PR #34, opened 2019-07-03
Minitaur environment

diegoferigo:
This PR adds a new environment to simulate the Minitaur robot in Ignition Gazebo. The vendored model was taken from [KodlabPenn/ghost_gazebo](https://github.com/KodlabPenn/ghost_gazebo).

The status of the PR is still exploratory, and currently I am still facing few issues:

- [ ] Problems with the five-bar linkage leg mechanism. The parallel mechanism is not working and I am not sure if the problem comes from Ignition Gazebo or Dart.
- [ ] Despite the models have been hardly tuned on Classic Gazebo, enabling a simple low-level PID joint controller makes the robot explode.

I would like to have a fixed-based environment to tune the models without dealing with contacts. I still didn't manage to make it work in Ignition Gazebo even following a similar structure of our [fixed icub models](https://github.com/robotology/icub-gazebo-wholebody/blob/master/models/icub_fixed_no_hands/icub_fixed_no_hands.sdf).


diegoferigo:
Moving temporarily this PR in the backlog. When I did the first experiments with the Minitaur sdf model included in this PR, I had problems with the simulation of the parallel legs in Ignition Gazebo. I am not really sure that parallel chains are currently supported.

Since the pybullet support is about to come, it might be easier to prototype everything with it and only then try to adapt the Task to work with Ignition Gazebo.

diegoferigo:
> I am not really sure that parallel chains are currently supported.

It seems that the dart plugin doesn't yet support closed chains, as reported in [this comment](https://bitbucket.org/ignitionrobotics/ign-physics/src/d3fd7dd3131792933a67124982776f7e500485bc/dartsim/src/FreeGroupFeatures.cc#lines-50:52):

> TODO(MXG): When the dartsim plugin supports closed-loop constraints, verify
    that this model is not attached to the world or any other models. If it's
    attached to anything external, then we should return an invalid identity.

diegoferigo:
I just opened [a dedicated upstream issue](https://bitbucket.org/ignitionrobotics/ign-physics/issues/25/support-of-closed-kinematic-chains).

diegoferigo:
Closing this PR since closed loop kinematic chains are not yet supported in Ignition Gazebo and there is no upstream progress on this missing feature.

PR #92, opened 2019-10-30
Test lgtm integration

diegoferigo:


diegoferigo:
I didn't find a way to add a new ppa in the docker image used by lgtm. The package `sudo` is installed but the user is not in the sudoers and cannot use it.

Anyway, Python is currently detected by lgtm analysis and has been already enabled. There are few warnings to fix, I will do it in another PR.

Closing this PR for the moment, maybe we'll find in the future a workaround for the ppa.

diegoferigo:
- Detail about how the automatic dependency system has been designed: https://blog.semmle.com/cplusplus-projects-lgtm/
- Statement that PPA cannot be added: https://lgtm.com/help/lgtm/analysis-faqs#whats-the-default-build-environment-for-cpp-projects

> There is no way to install packages from external repositories (PPAs) since that would allow privilege escalation.

PR #122, opened 2020-01-24
Update to IgnitionRobotics Citadel

diegoferigo:
For this time it's ok doing all these changes, but in the long they are not very maintainable. I'd like to find a smarter way to handle the major release versions of the ignition robotic suit.

### DIFF
--- a/.docker/Dockerfile.ci
+++ b/.docker/Dockerfile.ci
@@ -39,7 +39,7 @@ RUN apt-get update &&\
     rm -rf /var/lib/apt/lists/*
 
 # Install ignition robotics
-ARG ignition_codename="blueprint"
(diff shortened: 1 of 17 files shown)


diegoferigo:
I'm not really sure why tests are segfaulting (here the [valgrind analysis](https://github.com/robotology/gym-ignition/pull/122/checks?check_run_id=409423711#step:10:6882)). I just tried to execute the tests locally in the same docker image and everything runs smoothly :thinking:

diegoferigo:
Update, the only compiler that works is gcc7, which was the one I used in the execution of the previous comment.

All other compilers (gcc8, clang7, clang9) segfault. cc @traversaro

PR #140, opened 2020-02-12
Merge upstream modifications into vendored Physics system

diegoferigo:
Aligned the system up to:

https://bitbucket.org/ignitionrobotics/ign-gazebo/commits/160e2f963b7b596890d3fa757a5276de7c7060c5

PR #141, opened 2020-02-12
Extract last applied joint forces

diegoferigo:
This PR introduces the support of extracting the last-applied generalized force of joints.

There are few caveats. For instance if physics rate = agent rate = controller rate, everything work as expected. Though, if these rates are different, the extracted force reference could not be what expected. In fact, this method will **always return the force reference applied in the _last step_**.

Users should be aware of this behaviour when using this method.

This PR depends on #140.

### DIFF
--- a/gym_ignition/base/robot/robot_joints.py
+++ b/gym_ignition/base/robot/robot_joints.py
@@ -117,6 +117,28 @@ class RobotJoints(ABC):
         """
 
     @abstractmethod
+    def joint_force(self, joint_name: str) -> float:
+        """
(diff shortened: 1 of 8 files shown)


diegoferigo:
> Ack, even if I would document that jont_force is an effort/torque, as a 1 dof force may be surprising for most users.

This goes in the same direction of clearing out in gym-ignition the difference between joint and DoF. Before v1 we have to address it since it could cause misunderstandings. For the time being let's keep this ambiguity and defer the fix to a future PR.

PR #147, opened 2020-02-13
Update cartpole and pendulum to use models from gym-ignition-models

diegoferigo:


diegoferigo:
Good catch! Addressed in diff https://github.com/robotology/gym-ignition/compare/e3b4d34dee44a00bd0daed97cad8b64f67157300..84bfaa0a1f313f5aeeedea1a22ac4cd02520d7dc.

PR #166, opened 2020-03-31
New JointController plugin to implement Position and Velocity joint control modes

diegoferigo:
This is the first example of a model plugin that could be inserted during runtime, i.e. after inserting the model in the simulation.

The plugin runs as a System stepped by Gazebo at the rate the server was configured. In order to run the PIDs at a different rate, users could define a different period with:

https://github.com/robotology/gym-ignition/blob/e7c6b7d5f566f356de54a4cf15fc4015aa013090/cpp/scenario/gazebo/include/scenario/gazebo/Model.h#L101-L102

There is a single PID per joint that must be configured when either `JointControlModel::Position` or `JointControlMode::Velocity` are enabled.


diegoferigo:
Ready to be reviewed @traversaro

diegoferigo:
> we have any pratical idea of whether this system does run before or after the physics update?

Yours is an excellent question, and I don't have a ready answer. Though, I did consider it while designing the plugin. If you notice, the output of the PIDs is computed in the `JointController::PreUpdate` method. This method for sure is running before `Physics::Update`, so in this sense the execution should run flawless.

Something instead that I'm not sure is whether `JointController::PreUpdate` is called before `Physics::PreUpdate`. Right now `Physics::PreUpdate` is not doing much (and nothing related to PIDs), but we should keep this in mind if the logic of that physics callback will be used more extensively.

diegoferigo:
I leave as a reference the link to a PR hitting upstream with a different joint controller https://github.com/ignitionrobotics/ign-gazebo/pull/216/

PR #173, opened 2020-04-09
Experimental multi-world support

diegoferigo:
This PR introduces the support to insert in the simulator multiple independent worlds.

What's interesting is that each world is simulated in the same process but in different threads. Each thread contains a `SimulationRunner` with one of the worlds. Worlds can have different physics and contain a different number of models, but all of them must share the same server configuration (step size, real-time factor, number of steps per run).

This architectures potentially enables gym-ignition to fully exploit the CPU resources of the machine for sampling from multiple world in parallel. I will push some Python code in another PR, but there are severe upstream limitations that prevent using the multi-world feature. In fact, upstream's multi-world setup has [few bugs](https://bitbucket.org/ignitionrobotics/ign-gazebo/issues/18/support-for-multiple-worlds) that haven't been solved yet.

### DIFF
--- a/cpp/scenario/gazebo/include/scenario/gazebo/GazeboSimulator.h
+++ b/cpp/scenario/gazebo/include/scenario/gazebo/GazeboSimulator.h
@@ -61,8 +61,10 @@ public:
     bool pause();
     bool running() const;
 
-    bool loadSdfWorld(const std::string& worldFile,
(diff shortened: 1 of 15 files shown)


diegoferigo:
Merging to move forward. Let me know anyway if there are modifications that could be addressed in the final `refactoring` -> `devel` merging.

PR #177, opened 2020-04-19
Add Randomizers: model and physics

diegoferigo:
This PR complements #176 and adds the missing piece used in the new Python package: Randomizers.

Before, the combination of Task + Runtime was enough to make an environment. Now it's no longer the case. Take as reference the following:

- Updated [`launch_cartpole.py`](https://github.com/diegoferigo/gym-ignition/blob/03eef30a7e7ad7501df3675854886ce0dc978fc8/examples/python/launch_cartpole.py) example.
- [`CartPoleDiscreteBalancing`](https://github.com/diegoferigo/gym-ignition/blob/03eef30a7e/python/gym_ignition_environments/tasks/cartpole_discrete_balancing.py) task, same as before.
- Updated [`GazeboRuntime`](https://github.com/diegoferigo/gym-ignition/blob/03eef30a7e/python/gym_ignition/runtimes/gazebo_runtime.py) that works with the new ScenarI/O APIs.
- [Environment registration](https://github.com/diegoferigo/gym-ignition/blob/03eef30a7e7ad7501df3675854886ce0dc978fc8/python/gym_ignition_environments/__init__.py#L24-L32) same as before.

This structure is comparable to what we used to have before. However, the `env` obtained from the `gyn.make` factory, is not usable by default.

Randomizers are developed as environment wrappers, and they mainly extend the `gym.Env.reset` method. They will add the model used by the task in the world [1] and optionally perform one or more of the following steps:

- Randomize the model using the new `SDFRandomizer` (for example the dynamics parameters of the model - mass, inertia, joint friction, ... - and the friction with ground). See the test in 52d16f30361d216107317dc6ac0f5b8e44e82555 for an example.
- Randomize the physics (at the moment we only support loading DART and randomizing gravity, but as soon as ign-physics will have more back-ends, the used physics engine and other exposed parameters could be randomized).

Two randomizers examples, one that does not apply any randomization and another that fully randomizes what's currently supported, can be seen in 4d474153cb107e18b61d2b6841ad50d1ce8a4aa0.

[1] This was previously done by the runtime since it was the only setup we were supporting. Now the randomizer could add as many models as are required. For example, in manipulation environments they could reset the manipulator and the work area.

### DIFF
--- a/python/gym_ignition/randomizers/base/model.py
+++ b/python/gym_ignition/randomizers/base/model.py
@@ -1,0 +1,27 @@
+# Copyright (C) 2020 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+import abc
+
+
+class ModelRandomizer(abc.ABC):
+
+    @abc.abstractmethod
+    def randomize_model(self) -> str:
+        """
+        Randomize the model.
+
+        Return:
+            A string with the randomized model.
+        """
+        pass
+
+    def seed_model_randomizer(self, seed: int) -> None:
+        """
+        Seed the randomizer to ensure reproducibility.
+
+        Args:
+            seed: The seed number.
+        """
+        pass
--- a/python/gym_ignition/randomizers/base/physics.py
+++ b/python/gym_ignition/randomizers/base/physics.py
@@ -45,7 +45,7 @@ class PhysicsRandomizer(abc.ABC):
         """
         pass
 
-    def add_rollout_to_physics(self) -> None:
+    def increase_rollout_counter(self) -> None:
         """
         Increase the rollouts counter.
         """
--- a/python/gym_ignition/randomizers/base/physics.py
+++ b/python/gym_ignition/randomizers/base/physics.py
@@ -1,0 +1,72 @@
+# Copyright (C) 2020 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+import abc
+from gym_ignition import scenario_bindings as bindings
+
+
+class PhysicsRandomizer(abc.ABC):
+    """
+    Abstract class that provides the machinery for randomizing physics in a Ignition
+    Gazebo simulation.
+
+    Args:
+        randomize_after_rollouts_num: defines after many rollouts physics should be
+        randomized (i.e. the amount of times :py:meth:`gym.Env.reset` is called).
+    """
+
+    def __init__(self, randomize_after_rollouts_num: int = 0):
+
+        self._rollout_counter = randomize_after_rollouts_num
+        self.randomize_after_rollouts_num = randomize_after_rollouts_num
+
+    @abc.abstractmethod
+    def randomize_physics(self, world: bindings.World) -> None:
+        """
+        Method that insert and configures the physics of a world.
+
+        By default this method loads a plugin that uses DART with no randomizations.
+        Randomizing physics engine parameters or changing physics engine backend could be
+        done by redefining this method and passing it to
+        :py:class:`~gym_ignition.runtimes.gazebo_runtime.GazeboRuntime`.
+
+        Args:
+            world: A world object without physics.
+        """
+        pass
+
+    def seed_physics_randomizer(self, seed: int) -> None:
+        """
+        Seed the randomizer to ensure reproducibility.
+
+        Args:
+            seed: The seed number.
+        """
+        pass
+
+    def add_rollout_to_physics(self) -> None:
+        """
+        Increase the rollouts counter.
+        """
+
+        if self.randomize_after_rollouts_num != 0:
+            assert self._rollout_counter != 0
+            self._rollout_counter -= 1
+
+    def physics_expired(self) -> bool:
+        """
+        Checks if the physics needs to be randomized.
+
+        Return:
+            True if the physics has expired, false otherwise.
+        """
+
+        if self.randomize_after_rollouts_num == 0:
+            return False
+
+        if self._rollout_counter == 0:
+            self._rollout_counter = self.randomize_after_rollouts_num
+            return True
+
+        return False
--- a/python/gym_ignition/randomizers/base/task.py
+++ b/python/gym_ignition/randomizers/base/task.py
@@ -1,0 +1,38 @@
+# Copyright (C) 2020 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+import abc
+from gym_ignition import base
+from gym_ignition import scenario_bindings as bindings
+
+
+class TaskRandomizer(abc.ABC):
+
+    @abc.abstractmethod
+    def randomize_task(self,
+                       task: base.task.Task,
+                       gazebo: bindings.GazeboSimulator,
+                       **kwargs) -> None:
+        """
+        Randomize a :py:class:`~gym_ignition.base.task.Task` instance.
+
+        Args:
+            task: the task to randomize.
+            gazebo: a :py:class:`~scenario_bindings.GazeboSimulator` instance.
+
+        Note:
+            Note the each task has a :py:attr:`~gym_ignition.base.task.Task.world`
+            property that provides access to the simulated
+            :py:class:`scenario_bindings.World`.
+        """
+        pass
+
+    def seed_task_randomizer(self, seed: int) -> None:
+        """
+        Seed the randomizer to ensure reproducibility.
+
+        Args:
+            seed: The seed number.
+        """
+        pass
--- a/python/gym_ignition/randomizers/base/task.py
+++ b/python/gym_ignition/randomizers/base/task.py
@@ -22,7 +22,7 @@ class TaskRandomizer(abc.ABC):
             gazebo: a :py:class:`~scenario_bindings.GazeboSimulator` instance.
 
         Note:
-            Note the each task has a :py:attr:`~gym_ignition.base.task.Task.world`
+            Note that each task has a :py:attr:`~gym_ignition.base.task.Task.world`
             property that provides access to the simulated
             :py:class:`scenario_bindings.World`.
         """
--- a/python/gym_ignition/randomizers/model/sdf.py
+++ b/python/gym_ignition/randomizers/model/sdf.py
@@ -1,0 +1,235 @@
+# Copyright (C) 2020 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+import numpy as np
+from lxml import etree
+from pathlib import Path
+from enum import auto, Enum
+from typing import Dict, List, NamedTuple, Union
+
+
+class Distribution(Enum):
+    Uniform = auto()
+    Gaussian = auto()
+
+
+class Method(Enum):
+    Absolute = auto()
+    Additive = auto()
+    Coefficient = auto()
+
+
+class RandomizationData(NamedTuple):
+    xpath: str
+    distribution: str
+    parameters: "DistributionParameters"
+    method: Method
+    ignore_zeros: bool = False
+    force_positive: bool = False
+    element: etree.Element = None
+
+
+class GaussianParams(NamedTuple):
+    variance: float
+    mean: float = None
+
+
+class UniformParams(NamedTuple):
+    low: float
+    high: float
+
+
+DistributionParameters = Union[UniformParams, GaussianParams]
+
+
+class RandomizationDataBuilder:
+
+    def __init__(self, randomizer: "SDFRandomizer"):
+
+        self.storage: Dict = {}
+        self.randomizer = randomizer
+
+    def at_xpath(self, xpath: str) -> "RandomizationDataBuilder":
+        self.storage["xpath"] = xpath
+        return self
+
+    def sampled_from(self,
+                     distribution: Distribution,
+                     parameters: DistributionParameters) -> "RandomizationDataBuilder":
+
+        self.storage["distribution"] = distribution
+        self.storage["parameters"] = parameters
+
+        if self.storage["distribution"] is Distribution.Gaussian and \
+            not isinstance(parameters, GaussianParams):
+            raise ValueError("Wrong parameters type")
+
+        if self.storage["distribution"] is Distribution.Uniform and \
+            not isinstance(parameters, UniformParams):
+            raise ValueError("Wrong parameters type")
+
+        return self
+
+    def method(self, method: Method) -> "RandomizationDataBuilder":
+
+        self.storage["method"] = method
+        return self
+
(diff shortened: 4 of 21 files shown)


diegoferigo:
The `SDFRandomizer` was designed after the discussion of https://github.com/robotology/gym-ignition/issues/41#issuecomment-612587703. It was inspired by https://github.com/ryanjulian/rllab/issues/61 and https://github.com/rlworkgroup/garage/pull/51. If you're curious about the used pattern, you can refer to [Builder Pattern](https://refactoring.guru/design-patterns/builder/python/example) that produces a [Fluent Interface](https://en.wikipedia.org/wiki/Fluent_interface#Python).

The mujoco randomization is somehow more complicated than what we can do with SDF. In fact, mujoco models store most of their parameters in the attributes of the DOM, instead SDF always use elements.

Here a list of additional features I implemented:

- The randomizer can operate on multiple elements matching a XPath. This means that, for example, with a single configuration entry we can randomize with the same statistics the masses of all the links.
- Novel `Method::Additive` to alter the variable with additive noise. This is useful in combination of a gaussian randomization since we can add noise with zero mean.
- Novel `force_positive` method. Depending on the selected randomization method and distribution, it could happen that a randomized value becomes less than zero. In order to avoid seeing dragons e.g. with negative masses, this method clips randomized values to force them being positive.

raffaello-camoriano:
> This structure is comparable to what we used to have before. However, the `env` obtained from the `gyn.make` factory, is not usable by default.

As far as I can tell from the above comment and from https://github.com/robotology/gym-ignition/commit/4d474153cb107e18b61d2b6841ad50d1ce8a4aa0#diff-64386c8ba5955b2a8256341650413e50, in a use case without any need for randomization the user would be required anyways to use an extension wrapper `GazeboEnvRandomizer` around the env produced by `gym.make`.
Is this correct? Would it be feasible to make the raw `gym.make` env compatible with the framework?

raffaello-camoriano:
> > in a use case without any need for randomization the user would be required anyways to use an extension wrapper GazeboEnvRandomizer around the env produced by gym.make.
> > Is this correct? Would it be feasible to make the raw gym.make env compatible with the framework?
> 
> It is correct and it was a consequence of the redesign. If you want the wrapped Task class to be independent from Gazebo, using a randomizer that just inserts the model and performs no randomizations is the only solution. In fact, the method `World.insertModel(urdf, pose, name)` is specific for Gazebo, and if used in the Task class it will make the task not work in a RealTimeRuntime.
> 
> However, in the case of a Task class that's supposed to work only with the Gazebo runtime (i.e. it does not target real-time transparency), populating the world in the `Task.reset_task` method is a great solution. In this case, a randomizer is not necessary. 

I see, thanks.

> I always consider generic tasks in my explanations since we're aiming to runtime transparency, but maybe for many users it is not a requirement that strict. We should keep it in mind when writing the documentation.

Agreed.

python/gym_ignition/randomizers/base/physics.py
```diff
@@ -0,0 +1,72 @@
...
+    def seed_physics_randomizer(self, seed: int) -> None:
+        """
+        Seed the randomizer to ensure reproducibility.
+
+        Args:
+            seed: The seed number.
+        """
+        pass
+
+    def add_rollout_to_physics(self) -> None:
```
raffaello-camoriano, line 48: This method name doesn't sound crystal clear to me.
What about `PhysicsRandomizer.increase_rollout_counter`?

python/gym_ignition/randomizers/base/model.py
```diff
@@ -0,0 +1,27 @@
...
+
+import abc
+
+
+class ModelRandomizer(abc.ABC):
+
+    @abc.abstractmethod
+    def randomize_model(self) -> str:
+        """
+        Randomize the model.
```
raffaello-camoriano, line 13: In the future documentation it would be very useful to have a list of all the randomizable quantities.

python/gym_ignition/randomizers/base/task.py
```diff
@@ -0,0 +1,38 @@
...
+                       **kwargs) -> None:
+        """
+        Randomize a :py:class:`~gym_ignition.base.task.Task` instance.
+
+        Args:
+            task: the task to randomize.
+            gazebo: a :py:class:`~scenario_bindings.GazeboSimulator` instance.
+
+        Note:
+            Note the each task has a :py:attr:`~gym_ignition.base.task.Task.world`
```
raffaello-camoriano, line 25: Note that

python/gym_ignition/randomizers/model/sdf.py
```diff
@@ -0,0 +1,235 @@
...
+            else:
+                raise ValueError("Method not recognized")
+
+        return etree.tostring(self._root, pretty_print=pretty_print).decode()
+
+    def new_randomization(self) -> RandomizationDataBuilder:
+        return RandomizationDataBuilder(randomizer=self)
+
+    def insert(self, randomization_data) -> None:
+        self._randomizations.append(randomization_data)
```
raffaello-camoriano, line 206: Why are the randomization configuration data appended? How is `self._randomization` organized? Is it a list with a `randomization_data` configuration object for each element to be randomized?

python/gym_ignition/randomizers/physics/dart.py
```diff
@@ -0,0 +1,29 @@
...
+
+from gym_ignition.randomizers.base import physics
+from gym_ignition import scenario_bindings as bindings
+
+
+class DART(physics.PhysicsRandomizer):
+    """
+    Class that configures the Ignition Gazebo physics with the DART physics engine and
+    no randomization.
+    """
```
raffaello-camoriano, line 13: Wouldn't it prevent confusion to call it e.g. `DART_no_rand` in view of other DART-targeted classes indeed introducing randomization?

python/gym_ignition/runtimes/gazebo_runtime.py
```diff
@@ -248,27 +261,10 @@ def world(self) -> bindings.World:
             if not ok_ground:
                 raise RuntimeError("Failed to insert the ground plane")
 
-        # Load the physics
-        self._load_physics(world)
+        # Load and randomize the physics
+        self.physics_randomizer.randomize_physics(world=world)
```
raffaello-camoriano, line 265: Neat

python/gym_ignition/randomizers/base/physics.py
diegoferigo: Yes that sounds better

python/gym_ignition/randomizers/base/model.py
diegoferigo: Given that you can access all the elements through XPath patterns, you can randomize everything in the SDF. The only constraint is that randomized quantities are elements and not attributes. This was a design choice that matches SDF, in fact all relevant data are store in elements.

python/gym_ignition/randomizers/model/sdf.py
```diff
@@ -0,0 +1,235 @@
...
+        if self.storage["distribution"] is Distribution.Uniform and \
+            not isinstance(parameters, UniformParams):
+            raise ValueError("Wrong parameters type")
+
+        return self
+
+    def method(self, method: Method) -> "RandomizationDataBuilder":
+
+        self.storage["method"] = method
+        return self
```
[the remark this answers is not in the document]
diegoferigo: Yes, the type hint should help in this case. I realized I didn't document yet this class, I will do it.

python/gym_ignition/randomizers/model/sdf.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def ignore_zeros(self, ignore_zeros: bool) -> "RandomizationDataBuilder":
+        self.storage["ignore_zeros"] = ignore_zeros
+        return self
```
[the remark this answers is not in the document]
diegoferigo: Here some documentation is needed (coming). The idea behind this configuration is to ignore elements that have a value of zero. Let's make an example.

Consider that you want to randomize the inertia matrix (ixx, iyy, ixy, ...). Usually only the main diagonal is non-zero. Using the `ignore_zeros` option you can enable the randomization on all of the 6 parameters, but if a value, let's say ixy, is zero, ixy is not randomized.

I think this is a feature that we need to write generic code that could be applied to multiple models. In fact, a very interesting feature that XPath enables is the application model-wise through a pattern.

python/gym_ignition/randomizers/model/sdf.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def add(self) -> None:
+
+        data = RandomizationData(**self.storage)
+
+        if len(self.randomizer.find_xpath(data.xpath)) == 0:
+            raise RuntimeError(f"Failed to find element matching XPath '{data.xpath}'")
+
+        self.randomizer.insert(randomization_data=data)
```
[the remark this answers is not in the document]
diegoferigo: You can learn more from the Builder pattern I linked in a comment above. `add` closes the chaining of the `*Builder` and insert itself into a buffer of the `SDFRandomizer` object before returning.

python/gym_ignition/randomizers/model/sdf.py
```diff
@@ -0,0 +1,235 @@
...
+        self._root: etree.Element = tree.getroot()
+
+        # List of randomizations
+        self._randomizations: List[RandomizationData] = []
+
+        # List of default values used with Method.Coefficient
+        self._default_values: Dict[etree.Element, float] = {}
+
+        # Store an independent RNG
+        self.rng = np.random.default_rng()
```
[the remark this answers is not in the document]
diegoferigo: The term RNG is widely known especially in CS and ML (even numpy uses it) but I can expand the comment if it makes it more clear.

python/gym_ignition/randomizers/model/sdf.py
diegoferigo: Tip: always check the type hinting, they're there to help you understand the type of the variables even in a language that is not statically typed.

You will find that `_randomizations` is a `List[RandomizationData]`. It is used in the `add` method of the `*Builder` to insert the new randomization in a buffer of this class.

python/gym_ignition/randomizers/physics/dart.py
diegoferigo: I thought about it and given that there could be multiple types of randomizations I decided to do the opposite. Call this just `DART` and use a longer and more descriptive name to classes that perform randomizations (e.g. `DARTRandomizeGravity`, ...).

python/gym_ignition/randomizers/base/physics.py
diegoferigo: Anyway, this is still a rough edge among all the randomizations. I managed to integrate the physics in the randomized quantities but the logic is more convoluted than model and task. This is currently working fine but I'm not fully satisfied of how it works. Though, I couldn't find yet a cleaner way to do it, but you can expect that it will improved in the future (hopefully, by removing this counter).

diegoferigo: Update in 3a8e360

python/gym_ignition/randomizers/base/task.py
diegoferigo: Done in af2c664

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
diegoferigo: Documented in ed0c1cb

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
diegoferigo: See documentation ed0c1cb

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
diegoferigo: See documentation ed0c1cb.

python/gym_ignition/randomizers/model/sdf.py
diegoferigo: See documentation ed0c1cb.

python/gym_ignition/randomizers/base/physics.py
raffaello-camoriano: Ok, souds good.

python/gym_ignition/randomizers/base/model.py
raffaello-camoriano: Ok, I see.

raffaello-camoriano: Users may find it useful to refer to a set of examples. The cartpole ones are a great point to start. In the future, it may be valuable to also provide an example involving a more complex robot like Panda or iCub.

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
raffaello-camoriano: Ok

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
raffaello-camoriano: Thanks for explaining this.

If `ignore_zeros == False`, then would also the out-of-diagonal zero elements be randomized?

If so, then we should basically always set `ignore_zeros == True` for all matrix quantities with zero elements by construction, right?

A possible future improvement for added flexibility in randomization scope definition, the `ignore_zeros` may also be defined as a binary mask over the matrix.

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
raffaello-camoriano: Ok, thanks.

python/gym_ignition/randomizers/base/model.py
diegoferigo: This PR introduces a test with a Panda:

https://github.com/robotology/gym-ignition/blob/b1ff4419743c71e49e457b42213170a6c0f2f415/tests/test_gym_ignition/test_sdf_randomizer.py#L215

 In general, tests are a great piece of implicit documentation that many ignore.

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
diegoferigo: Not really. See the following as reference:

https://github.com/robotology/gym-ignition/blob/b1ff4419743c71e49e457b42213170a6c0f2f415/tests/test_gym_ignition/test_sdf_randomizer.py#L284-L291

The XPath `*/joint/axis/dynamics/damping` will match with the viscous friction element of all joint. If a joint has zero friction by default, the randomization is not applied on it.

python/gym_ignition/randomizers/model/sdf.py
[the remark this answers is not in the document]
raffaello-camoriano: It is frequent to come across similar acronyms and namings with different meanings across subfields and it shouldn't be taken for granted that an acronyms is self-explanatory. In general, defining acronyms at least once in the documentation (or in a paper) is always a good practice.

python/gym_ignition/randomizers/physics/dart.py
raffaello-camoriano: Ok

python/gym_ignition/randomizers/model/sdf.py
raffaello-camoriano: Ok, thanks.

PR #179, opened 2020-04-20
Extract extended contact data (force, normal, penetration depth)

diegoferigo:
This PR shows how https://github.com/ignitionrobotics/ign-physics/pull/40 could be used to extract data provided by the DART ign-physics plugin.

Still in WIP state waiting to receive feedback from upstream.

### DIFF
--- a/tests/test_scenario/test_contacts.py
+++ b/tests/test_scenario/test_contacts.py
@@ -1,0 +1,226 @@
+# Copyright (C) 2020 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+import pytest
+pytestmark = pytest.mark.scenario
+
+import numpy as np
+from ..common import utils
+import gym_ignition_models
+from typing import Callable
+from gym_ignition.utils import misc
+from ..common.utils import gazebo_fixture as gazebo
+from gym_ignition import scenario_bindings as bindings
+
+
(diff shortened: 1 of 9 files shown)


diegoferigo:
@traversaro this PR is ready to be reviewed even though it all depends on how the upstream PR will be handled. Can you please double check the method `Link::contactWrench()`? Tests on my machine pass and it should be fine, but all of them use a planar surface.

tests/test_scenario/test_contacts.py
```diff
@@ -0,0 +1,226 @@
...
+    assert len(cube3.contacts()) == 2
+
+    contact1 = cube3.contacts()[0]
+    contact2 = cube3.contacts()[1]
+
+    assert contact1.bodyA == "cube3::cube"
+    assert contact2.bodyA == "cube3::cube"
+
+    assert contact1.bodyB == "cube1::cube"
+    assert contact2.bodyB == "cube2::cube"
```
raffaello-camoriano, line 206: Is cube1::cube always associated with contact1 (and cube2 with contact2)?

diegoferigo: Since the simulation is reproducible this order is always respected. Though, I had to check beforehand because in cases like this in which contacts are created in the same moment you cannot know a priory how contacts get detected.

diegoferigo:
Damn, I forgot to change the target branch to `devel` before merging `refactoring`. Now github does not allow to me change the targeted branch. I guess I have to reopen a new PR. Let's wait some upstream activity before doing it.

PR #184, opened 2020-04-27
Fix controllers after model removal

diegoferigo:


diegoferigo:
Yes I agree, but this is a necessary workaround. The real solution is unloading the plugin on model removal. I opened https://github.com/ignitionrobotics/ign-gazebo/issues/113 (the previous issue where we discussed has been closed since it was not directly related).

PR #193, opened 2020-05-01
Generate idiomatic undercase Python bindings

diegoferigo:
Refer to http://www.swig.org/Doc4.0/SWIG.html#SWIG_advanced_renaming for more details.


diegoferigo:
Thanks for the link to the proposal, it's worth mentioning.

PR #220, opened 2020-08-14
ScenarI/O: real C++ interfaces and standalone Python package

diegoferigo:
:warning: This is a meta pull requests that collects a bigger set of other pull requests. Consider to review the individual PRs to get smaller and more readable diffs.

This PR aims to the following two objectives:

1. Convert the ScenarI/O APIs, currently only implemented for Ignition Gazebo, to real C++ interfaces. The current Ignition Gazebo classes will become one of the backends we'll support.
1. Make the `scenario` Python package standalone. Currently there's a complex relation between `gym_ignition` and `scenario`, where the latter are imported from the former in downstream code. The final result will be:
    - **`scenario`** will become a standalone package, and the existing Gazebo implementation can be imported with `import scenario.bindings.gazebo`. It should setup the environment without the need of `gym_ignition` (including dlopen workarounds and exporting env vars).
   - **`gym_ignition`** will become a pure Python package that just imports `scenario` and uses it to provide helper classes to write cleaner OpenAI Gym code for robotics. No more no less.

---

If you're aware of the [pytorch-lightning](https://github.com/PyTorchLightning/pytorch-lightning) project, I can say that:

`scenario : gym_ignition = pytorch : pytorch-lightning`

---

In this PR, the entire task will get merged chunk after chunk. Here below the expected sequence:

- [x] Create in the `scenario::core` namespace the new ScenarI/O interfaces as subset of the current Gazebo implentation https://github.com/robotology/gym-ignition/pull/219
- [x] Update the Gazebo implementation to inherit from the new interfaces #221 
- [x] Update the rest of the `scenario::gazebo` namespace to use the new `scenario::core` resources #222 
- [x] Update the CMake project of the SWIG bindings to create a real `scenario` package splitting`scenario.core` and `scenario.gazebo` in two different subpackages #223
- [x] Update the Python project (`gym_ignition`, `gym_ignition_environments`) #224 
- [x] Fix the tests (until here CI won't be green) #225 

In particular, beyond the main objectives of this PR, I'd like to finally provide a decent base of documentation that so far has been lacking. It won't be perfect nor 100% complete, but it's at least a start.


diegoferigo:
cc @traversaro @paolo-viceconte @raffaello-camoriano @anqingd @CarlottaSartore @DanielePucci

PR #231, opened 2020-08-18
Simplify controlling models with PIDs

diegoferigo:
`core.JointControlMode_position` is among the supported joint control modes we have implemented. Given a model, it can be enabled model-wise as follows:

```python
# Enable position control
panda: scenario.bindings.gazebo.Model = world.get_model("panda").to_gazebo()
panda.set_controller_period(0.001)
panda.set_joint_control_mode(core.JointControlMode_position)

# Set a reference
joint1 = panda.get_joint("panda_joint1")
joint1.set_position_target(position=joint1_reference)
```

However, this was not enough. In order to insert the PIDs in the simulation, the user also had to add the [`JointController`](https://github.com/robotology/gym-ignition/tree/devel/cpp/scenario/plugins/JointController) plugin to the model as follows:

```python
panda.insert_model_plugin("JointController", "scenario::plugins::gazebo::JointController")
```

The main reason is that, similarly to custom controllers (#167), also the JointController is a plugin that runs in its own thread. This was done in this way to allow running Python code at a lower rate than the controller and the physics (by executing multiple iterations every `GazeboSimulator.run`).

However, this plugin insertion is quite a low-level detail, and it requires the user to downcast the `scenario.bindings.core.Model` object to a `scenario.bindings.gazebo.Model`. This PR loads the plugin automatically as soon as either Position or Velocity control modes are enabled to any of the model joints.


diegoferigo:
cc @paolo-viceconte

PR #233, opened 2020-08-20
Refactor CMake project

diegoferigo:
This PR cleans up and refactors the entire CMake project. Sorry for the big diff, it required many hours and basically I had to edit pretty much all the `CMakeLists.txt`, there was no simple sequence of commits I could do.

Big changes:

- The differences between the `PyPI` build type and the canonical ones are much reduced (i.e. there are way less `if(${CMAKE_BUILD_TYPE} STREQUAL "PyPI")`)
- Full CMake install support with new COMPONENTS, importing the targets is now easy and modular :tada:
- ... and also the PyPI package now ships all the headers and CMake files to import the project!

The only quite dirty part of this refactoring is how the `tiny-process-lib` is handled. I had a lot of troubles while I was trying to avoid installing it. I initially tried to force it being STATIC, but also STATIC libraries have to be installed when linked against other STATIC targets. So I went towards the OBJECT type. If you're interested in the topic, have a look to `deps/CMakeLists.txt` and `cpp/scenario/gazebo/CMakeLists.txt`.


diegoferigo:
@traversaro I asked your review but I know you're AFK also this week. If you don't have time in the next few days, I'll merge this PR without your review. If you have some time when you'll be back, I'd appreciate a feedback, and I will address your comments in a new PR.

Issue #240, opened 2020-09-08
Prepare v1.0 release


diegoferigo:
This issue recaps all the tasks remaining for the first stable release of gym-ignition. The lock-down was a very prolific coding period, and since we arrived to the letter C after Aloha and Biro, the codename couldn't be anything else than Corona :crown:

- [x] Wait the next binary release of ign-gazebo. Hopefully it will solve https://github.com/dic-iit/gazebo-scenario-plugins/pull/2#issuecomment-684539294 (#251).
- [x] Officially update the minimum Python version to 3.8 (https://github.com/robotology/gym-ignition/pull/255).
- [x] Provide helpers to insert plugins and their context similar to [`plugin.py`](https://github.com/dic-iit/gazebo-scenario-plugins/blob/master/python/gazebo_scenario_plugins/plugin.py). We should replace for example in [`computed_torque_fixed_base.py`](https://github.com/robotology/gym-ignition/blob/devel/python/gym_ignition/controllers/gazebo/computed_torque_fixed_base.py) the Nametuple with dataclass (https://github.com/robotology/gym-ignition/pull/264).
- [ ] Merge the examples developed in the [`IFI/sprint`](https://github.com/robotology/gym-ignition/tree/IFI/sprint) branch.
- [x] Fix the demo section of the readme #238 (https://github.com/robotology/gym-ignition/pull/270).
- [x] Improve the README (https://github.com/robotology/gym-ignition/pull/270)
- [x] Add a `Panda` class to `gym_ignition_environments` even if there are no tasks yet. It can be useful in the examples and tests (#265).
- [x] Install iDynTree from pip and load the CMake project from there (https://github.com/robotology/gym-ignition/pull/255)
- [x] PEP517 and PEP518 compliance (https://github.com/robotology/gym-ignition/pull/255)
- [x] Allow installing Python bindings in a custom location (https://github.com/robotology/gym-ignition/pull/263)
- [x] Fix type hints of `Tuple[MyClass, ...]` removing the dots (https://github.com/robotology/gym-ignition/pull/268)
- [x] Make self-collisions a gazebo-only API (https://github.com/robotology/gym-ignition/pull/268)

Possible additions:

- [x] Updated non-linear IK class for both fixed and floating-base robots based on iDynTree (https://github.com/robotology/gym-ignition/pull/266)
- [x] New generic class for rigid-body dynamics based on `iDynTree::KinDynComputations` (https://github.com/robotology/gym-ignition/pull/266)

[...]

PR #257, opened 2020-10-30
Unify CI and CD workflows

diegoferigo:
https://github.com/robotology/gym-ignition/pull/251 updated the CI workflow removing the usage of docker images. 

While docker images are very convenient to ship all the dependencies, they create difficulties in dealing with the cases where a new PR updates the docker images and also needs them for CI to pass. This is not possible because the images are pushed only when the PR is merged. Therefore, all these kind of situation had to be split in two parts, or a manual push of images was necessary. This is quite a complex approach, that only who designed the CI/CD pipeline is fully aware of (me :smile:). In order to simplify others to contribute and, maybe, maintain in the future this project, I decided to rationalize it by removing the usage of docker images also for the PyPI workflow.

Then, I realized after #251, that makes us compatible with upstream, and the new PyPI support of iDynTree, that provides a static importable project, the CI pipeline is equivalent to the former PyPI pipeline. There's no need to handle them differently, forcing a static compilation of the dependencies when packaging.

This means that now we can deploy the Python packages generated and tested by the CI pipeline. 

Beyond the unification, this PR fixes how release event are handled, the previous PyPI workflow had few bugs. This PR is in preparation of #240.


traversaro (approved):
Yes, relying on (custom) Docker images for CI is not ideal, I think we experienced this in several projects.

PR #290, opened 2021-02-09
Force loading tensorflow shared libraries to prevent segfault

diegoferigo:
The segfault is a know problem and it is already documented in the FAQs:

https://github.com/robotology/gym-ignition/blob/321df6a2f93da829682ead2f428022ec1c4ef416/docs/sphinx/info/faq.rst#L4-L18

This PR is an attempt to prevent the occurrence, similar to https://github.com/apache/arrow/pull/2210. Additional information can be found in https://github.com/robotology/gym-ignition/issues/279#issuecomment-774730540.

Note that loading only the shared libraries is much faster than loading the entire `tensorflow` package.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -130,7 +130,7 @@ jobs:
         run: |
           apt-get update
           apt-get install -y --no-install-recommends \
-            python3-numpy libxml2-dev coinor-libipopt-dev libeigen3-dev libassimp-dev swig
+            libxml2-dev coinor-libipopt-dev libeigen3-dev libassimp-dev swig
 
       # =============
       # Build project
--- a/bindings/__init__.py
+++ b/bindings/__init__.py
@@ -82,7 +82,8 @@ def import_gazebo() -> None:
 
     # Preload the shared libraries of tensorflow if the package is installed.
     # If tensorflow is imported after scenario.bindings.gazebo, the application segfaults.
-    preload_tensorflow_shared_libraries()
+    if os.environ.get("SCENARIO_DISABLE_TENSORFLOW_PRELOAD") != "1":
+        preload_tensorflow_shared_libraries()
 
     # Import SWIG bindings
     # See https://github.com/robotology/gym-ignition/issues/7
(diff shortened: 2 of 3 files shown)


traversaro (approved):
Ok for me. Do you know if there is any python way for permit users to opt out/disable this workaround?

diegoferigo:
I think we can use an ad-hoc environment variable, such as `SCENARIO_DISABLE_TENSORFLOW_PRELOAD`. It's a good idea having it in place.

diegoferigo:
Ouch, something's wrong [in CI](https://github.com/robotology/gym-ignition/pull/290/checks?check_run_id=1863190713#step:13:34) with idyntree and numpy:

```
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/idyntree/__init__.py", line 1, in <module>
    from . import swig
  File "/usr/local/lib/python3.8/dist-packages/idyntree/swig.py", line 13, in <module>
    from . import _iDynTree
ImportError: numpy.core.multiarray failed to import
```

diegoferigo:
The CI is building the wheel from the `devel` branch of idyntree. The recent https://github.com/robotology/idyntree/pull/822 (cc @GiulioRomualdi) introduced a new variant of `KinDynComputations::setCurrentRobotConfiguration`, and in this context SWIG removes the keyword arguments from the bindings.

cba57f6 should solve the problem.

GiulioRomualdi:
> CI is still failing, this time in the IK test.
> 
> The CI is building the wheel from the `devel` branch of idyntree. The recent [robotology/idyntree#822](https://github.com/robotology/idyntree/pull/822) (cc @GiulioRomualdi) introduced a new variant of `KinDynComputations::setCurrentRobotConfiguration`, and in this context SWIG removes the keyword arguments from the bindings.
> 
> [cba57f6](https://github.com/robotology/gym-ignition/commit/cba57f6200ffe58ea8b31e429dec24885ff017bf) should solve the problem.

I'm sorry 😔

diegoferigo:
@GiulioRomualdi Using kwargs is not mandatory, but it's useful to provide more context that often would require referring to the documentation. The price to pay is a more tight alignment with upstream that occasionally triggers minor problems like this one. No problem at all ;) Actually the span-based methods could be quite useful also to Python, but I fear that for exploiting them the most, pybind11 is necessary (and here we're based on SWIG).

PR #311, opened 2021-03-29
New Feature: Support raw SDF strings in insertModel

FirefoxMetzger:
This PR adds the ability to pass a raw sdf string into `world.insertModel`. The way this works is that it first tries to parse the input string as a raw SDF string (via `utils::getSdfRootFromString`) and if that fails it continues an tries to parse the input string as a file path.

Potentially, this could be done the other way around, giving priority to paths. I was worried, however, that we might encounter weird bugs if a humongous string (entire worlds) gets passed into a function that expects a file path. Hence it is currently the other way around.

Here is an example test case that spawns a cube in a random color for 5 seconds.

```python
from scenario import gazebo as scenario_gazebo
import random
import time


sdf_string = """
    <?xml version="1.0" ?>
    <sdf version="1.6">
        <model name='box'>
        <pose>0 0 0.5 0 -0 0</pose>
            <link name='box_link'>
                <inertial>
                <inertia>
                    <ixx>1</ixx>
                    <ixy>0</ixy>
                    <ixz>0</ixz>
                    <iyy>1</iyy>
                    <iyz>0</iyz>
                    <izz>1</izz>
                </inertia>
                <mass>1</mass>
                </inertial>
                <collision name='box_collision'>
                <geometry>
                    <box>
                        <size>1 1 1</size>
                    </box>
                </geometry>
                <surface>
                    <friction>
                        <ode/>
                    </friction>
                    <contact/>
                </surface>
                </collision>
                <visual name='box_visual'>
                <geometry>
                    <box>
                    <size>1 1 1</size>
                    </box>
                </geometry>
                <material>
                    <ambient>1 0 0 1</ambient>
                    <diffuse>{diffuse}</diffuse>
                    <specular>1 0 0 1</specular>
                </material>
                <plugin name='__default__' filename='__default__'/>
                </visual>
            </link>
            <plugin name='__default__' filename='__default__'/>
        </model>
    </sdf>
"""


scenario_gazebo.set_verbosity(scenario_gazebo.Verbosity_error)
gazebo = scenario_gazebo.GazeboSimulator(step_size=0.001, rtf=1.0, steps_per_run=1)
gazebo.initialize()
world = gazebo.get_world()

colors = [
    "1 0 0 1",
    "0 1 0 1",
    "0 0 1 1",
    "1 0 1 1",
    "1 1 0 1",
    "1 1 1 1",
    "0 1 0 1",
    "0 1 1 1",
]
assert world.insert_model(sdf_string.format(diffuse=random.choice(colors)))
gazebo.gui()
gazebo.run(paused=True)
time.sleep(5)
gazebo.close()

```

I haven't tested this yet, but theoretically, this should work seamlessly in combination with my Fuel PR, as the Fuel callback is registered globally. This means that Fuel models can also be inserted via something like

```
gazebo.get_world().insertModel(
"""
    <?xml version="1.0" ?>
    <sdf version="1.6">
        <include>
          <uri>https://fuel.ignitionrobotics.org/1.0/OpenRobotics/models/Table</uri>
          <name>table1</name>
          <pose>0 0 0 0 -0 1.5708</pose>
        </include>
    </sdf>
"""
)
```

### DIFF
--- a/cpp/scenario/gazebo/src/World.cpp
+++ b/cpp/scenario/gazebo/src/World.cpp
@@ -278,12 +278,16 @@ scenario::core::ModelPtr World::getModel(const std::string& modelName) const
     return pImpl->models[modelName];
 }
 
-bool World::insertModel(const std::string& modelFile,
+bool World::insertModel(const std::string& modelString,
                         const core::Pose& pose,
                         const std::string& overrideModelName)
 {
-    const std::shared_ptr<sdf::Root> modelSdfRoot =
-        utils::getSdfRootFromFile(modelFile);
+    std::shared_ptr<sdf::Root> modelSdfRoot;
+
+    modelSdfRoot = utils::getSdfRootFromString(modelString);
+
+    if (!modelSdfRoot)
+        modelSdfRoot = utils::getSdfRootFromFile(modelString);
 
     if (!modelSdfRoot) {
         return false;
--- a/cpp/scenario/gazebo/src/World.cpp
+++ b/cpp/scenario/gazebo/src/World.cpp
@@ -86,7 +86,7 @@ public:
 
         // Get the final name of the model
         if (overrideModelName.empty()) {
-            assert(modelSdfRoot.ModelByIndex(ModelIndex));
+            assert(modelSdfRoot->ModelByIndex(ModelIndex));
             finalModelEntityName =
                 modelSdfRoot->ModelByIndex(ModelIndex)->Name();
         }
--- a/cpp/scenario/gazebo/src/World.cpp
+++ b/cpp/scenario/gazebo/src/World.cpp
@@ -65,6 +65,116 @@ public:
     {
         std::vector<std::string> modelNames;
     } buffers;
+
+public:
+    bool insertModel(const std::shared_ptr<sdf::Root>& modelSdfRoot, const core::Pose& pose,
+                        const std::string& overrideModelName, World& world)
+    {
+        if (modelSdfRoot->ModelCount() != 1) {
+            sError << "The SDF file contains more than one model" << std::endl;
+            return false;
+        }
+
+        constexpr size_t ModelIndex = 0;
+
+        // Every SDF model has a name. In order to insert multiple models from
+        // the same SDF file, the modelData struct allows providing a scoped
+        // name.
+        std::string finalModelEntityName;
+
+        // Get the final name of the model
+        if (overrideModelName.empty()) {
+            assert(modelSdfRoot.ModelByIndex(ModelIndex));
+            finalModelEntityName =
+                modelSdfRoot->ModelByIndex(ModelIndex)->Name();
+        }
+        else {
+            finalModelEntityName = overrideModelName;
+        }
+
+        // Check for model name clash
+        const std::vector<std::string>& modelNames = world.modelNames();
+        if (std::find(
+                modelNames.begin(), modelNames.end(), finalModelEntityName)
+            != modelNames.end()) {
+            sError << "Failed to insert model '" << finalModelEntityName
+                   << "'. Another entity with the same name already exists."
+                   << std::endl;
+            return false;
+        }
+
+        // Rename the model.
+        // NOTE: The following is not enough because the name is not serialized
+        // to
+        //       string. We need also to operate directly on the raw element.
+        const_cast<sdf::Model*>(modelSdfRoot->ModelByIndex(ModelIndex))
+            ->SetName(finalModelEntityName);
+
+        // Update the name in the sdf model. This is necessary because model
+        // plugins are loaded right before the creation of the model entity and,
+        // instead of receiving the model entity name, they receive the model
+        // sdf name.
+        if (!utils::renameSDFModel(
+                *modelSdfRoot, finalModelEntityName, ModelIndex)) {
+            sError << "Failed to rename SDF model" << std::endl;
+            return false;
+        }
+
+        if (utils::verboseFromEnvironment()) {
+            sDebug << "Inserting a model from the following SDF:" << std::endl;
+            std::cout << modelSdfRoot->Element()->ToString("") << std::endl;
+        }
+
+        // Create the model entity
+        ignition::gazebo::Entity modelEntity;
+        modelEntity = this->sdfEntityCreator->CreateEntities(
+            modelSdfRoot->ModelByIndex(ModelIndex));
+
+        // Attach the model entity to the world entity
+        this->sdfEntityCreator->SetParent(modelEntity, world.m_entity);
+
+        {
+            // Check that the model name is correct
+            assert(modelSdfRoot->ModelCount() == 1);
+            std::string modelNameSDF =
+                modelSdfRoot->ModelByIndex(ModelIndex)->Name();
+            std::string modelNameEntity = utils::getExistingComponentData< //
+                ignition::gazebo::components::Name>(world.m_ecm, modelEntity);
+            assert(modelNameSDF == modelNameEntity);
+        }
+
+        // Create the model
+        auto model = std::make_shared<scenario::gazebo::Model>();
+
+        // Initialize the model
+        if (!model->initialize(modelEntity, world.m_ecm, world.m_eventManager)) {
+            sError << "Failed to initialize the model" << std::endl;
+            if (!world.removeModel(finalModelEntityName)) {
+                sError << "Failed to remove temporary model after failure"
+                       << std::endl;
+            }
+            return false;
+        }
+
+        // Create required model resources. This call prepares all the necessary
+        // components in the ECM to make our bindings work.
+        if (!model->createECMResources()) {
+            sError << "Failed to initialize ECM model resources" << std::endl;
+            return false;
+        }
+
+        // Set the initial model pose.
+        // We directly override the Pose component instead of using
+        // Model::resetBasePose because it would just store a pose command that
+        // needs to be processed by the Physics system. Overriding the
+        // component, instead, has direct effect.
+        if (pose != core::Pose::Identity()) {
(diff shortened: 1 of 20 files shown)


cpp/scenario/gazebo/src/World.cpp
```diff
@@ -278,12 +278,16 @@ scenario::core::ModelPtr World::getModel(const std::string& modelName) const
     return pImpl->models[modelName];
 }
 
-bool World::insertModel(const std::string& modelFile,
+bool World::insertModel(const std::string& modelString,
```
diegoferigo, line 281: We've got lucky here that SWIG does not generate `kwargs` for C++ methods with default arguments like this one, otherwise this change would break Python code. Nonetheless, I'm am debating whether a more explicit `World::insertModelFromString()` method would be preferable.

diegoferigo (changes requested):
The PR LGTM apart from the discussion about creating a new method. Could you please add a third cube in [this test](https://github.com/robotology/gym-ignition/blob/bba62416e97fe5788019e32cae66355bcb9fe70a/tests/test_scenario/test_world.py#L124) loaded from string? You can get the string from the helper below:

https://github.com/robotology/gym-ignition/blob/bba62416e97fe5788019e32cae66355bcb9fe70a/tests/common/utils.py#L100

cpp/scenario/gazebo/src/World.cpp
FirefoxMetzger: Good catch! In a pure-python API this would indeed be non-backward compatible. Then again, we could easily fix it by re-introducing `modelFile` as a kwarg and plumbing it with a deprecation warning.

It might be better, yes. I'm not sure about the mismatch between `insertModelFromString` and `insertModel`; perhaps `insertModelFromString` and `insertModelFromFile`? 

In pure C++ I would have liked to see an overload of the form 

```
bool World::insertModel(const std::string& sdf_string, ...
bool World::insertModel(const std::filesystem::path& sdf_file, ...
```

Unfortunately, `sdf_file` is also a string making it effectively the same function.

In Python, I would have opted for a signature of the form `world.insert_model(self, model_sdf)` where `model_sdf` is an sdf-like object, aka, everything that can be meaningfully converted into sdf. This way, the same signature can cover a large array of inputs, from raw strings, over file paths, to more complex things like streams or fuel URLs (which we can wrap into a minimal sdf and load using the fuel loader, absorbing `getModelFileFromFuel`). This seems cleaner than having a new function for each input type. 

In either case, I'm happy to change, just let me know :)

FirefoxMetzger:
A side-effect of the current implementation is that loading models from file currently generates an error log message of the form

```
Error [parser.cc:496] Error parsing XML from string: Error=XML_ERROR_PARSING_TEXT ErrorID=10 (0xa) Line number=1
[Err] [helpers.cpp:79] Failed to load sdf string
[Err] [helpers.cpp:82] Error Code 27 Msg: Unable to SDF string: [path/to/file.sdf]
```

since it tries to parse the string as a file and fails. I can update the log messages, or we can use a different signature; let me know.

Also, @diegoferigo should I change the target of this PR to devel?

cpp/scenario/gazebo/src/World.cpp
diegoferigo: > perhaps `insertModelFromString` and `insertModelFromFile`?

I like the propolal of these explicit methods. In order not to break the APIs, maybe what could be done is providing both new specific methods `*{FromString,FromFile}` and in future `*FromFuel` (I like the idea to move it here), and an overall method `insertModel(std::string, core::Pose, std::string)`  that tries to figure out automatically what the user wants. In edge cases, automatic detection could fail, in this way users can fall back to specific methods.

We can split multiple PRs:

1. This one: move [this code](https://github.com/robotology/gym-ignition/blob/def00f7e47c8ab8e117b2bf77056226501bb7e68/cpp/scenario/gazebo/src/World.cpp#L292-L388) to a new:
   ```cpp
   World::Impl::insertModel(
       const std::string& modelFile, 
       const core::Pose& pose, 
       const std::string& overrideModelName,
       const World& world)  // This is needed just for this->modelNames()
    ```
   and call it from both the existing `World::insertModel` (just from file for the moment) and the new `World::insertModelFromString`
2. Next PR could add a new `World::insertModelFromFile` / `World::insertModelFromFuel` and implement the automatic detection. I remember in the past having problems in multi-os and multi-compiler with `std::filesystem`, few compilers were not happy with their partial implementation, but it was early times. Maybe with the compilers we officially support after the switch to Focal, the situation has improved. Worse case scenario, also `ignition/common/Filesystem.hh` could be used. 

Yesterday I had a look at Edifice and the sdformat11 deprecations, now `sdf::Root` is limited to contain just one `sdf::Model`, therefore the logic of these methods will be simplified with the next release.

Last comment, last week I started thinking about [frame semantics](http://sdformat.org/tutorials?tut=pose_frame_semantics_proposal), and particularly inserting models using `<pose relative_to='frame_name'>0 0 0 0 0 0</pose>`. I originally thought that it could have affected the signature of these `insertModel*` methods, but thinking again now, by extending `core::Pose` maybe we can prevent that and live happy without adding even more methods or breaking existing symbols.

diegoferigo:
> A side-effect of the current implementation is that loading models from file currently generates an error log message of the form
> 
> [...]
> 
> since it tries to parse the string as a file and fails. I can update the log messages, or we can use a different signature; let me know.

I realized it yesterday and tried to think to some workaround, but it seems to me that moving the error messages out from `utils.cpp` would be necessary. Unfortunately there are multiple locations where we assume that are those methods the responsible of creating the error messages,

I slept on the idea of the introduction of more explicit methods for inserting models, you can find it in https://github.com/robotology/gym-ignition/pull/311#discussion_r603885192. This logging problem should not occur in this case.  

> Also, @diegoferigo should I change the target of this PR to devel?

Yes, please update the branch.

cpp/scenario/gazebo/src/World.cpp
diegoferigo: > For this PR I would suggest to add both World::insertModelFromFile and World::insertModelFromString. These can exist in parallel with the old World::insertModel without breaking backwards compatibility.

In this case, `World::insertModel` can become a relay of `World::insertModelFromFile`. If you want to integrate it in this PR, ok from my side. Otherwise I can do it in a new one.

> In the second PR we can choose between two options: (1) Depreciate insertModel and only have explicit functions and rely on the user explicitly specifying the conversion. (2) Make insertModel auto-detect the appropriate conversion to sdf (perhaps with a method="auto" overwrite option) and depreciate the explicit methods. It's a philosophical decision on whether or not it makes sense to force the user to be explicit. Depreciation could happen with Fortress and would be in line with the current "under rapid development" label of the library. We could of course keep both, but the redundancy might be confusing to a user that tries to figure out which method to use.

I'd go towards (2), but it's not urgent, Fortress in any case is far away.

> In a third PR we can address World::insertModelFromFuel(std::str url, ...) (or the ability of understanding fuel URLs as input to insertModel) in favor of the current loadModelFromFuel+insertModel. Note that fuel support is already implied if it occurs as an include inside of models loaded from *fromString or *fromFile as soon as #309 is merged.

Yep, what I meant is providing a way into `GazeboSimulator` to do something similar to what today is done by  `utils::getModelFileFromFuel`.

> relative_to is something that we should investigate, too. Though, I think this will be a separate PR and I would cross that bridge once we get there.

Definitely. From my side it does not have a great priority, even if it'd be nice to have.

FirefoxMetzger: > move this code to a new `World::Impl::insertModel`

Actually, this doesn't work trivially, because `World::insertModel` calls `World::modelNames` which depends on attributes only available to `World` but not available to `World::Impl`, namely `ignition::gazebo::EntityComponentManager* m_ecm`.

Are you happy with me creating a private function inside of `World`, e.g. `World::insertModel(sdf::Root ...)` which acts as the shared base for the methods discussed above? If not, how would you like to have the code structured?

diegoferigo: A private method is a possible solution, even though I'm not yet sure I understood its necessity. I quickly tried locally, and the following signature inside `World::Impl` seems working fine:

```cpp
    bool insertModel(const std::shared_ptr<sdf::Root>& root,
                     const core::Pose& pose,
                     const std::string& overrideModelName,
                     World& world) // Note: this is a scenario::gazebo::World, 
                                   //       that has m_ecm attribute
```

Am I missing any detail?

FirefoxMetzger: Yep, doing this can also work. It just seems odd to pass the outer class into a nested inner class because the inner needs access to outer's attributes. Having a private method in outer is likely better/cleaner software engineering. 

It depends a bit on what the `*::Impl` inner classes are supposed to be. I haven't quite worked this out yet, which is why I asked how the code should be structured. Are they meant to be structs that capture data that belongs together, or are they a means of hiding private attributes and methods?

diegoferigo: I agree with you, what you described is the only drawback of the [pimpl idiom](https://cpppatterns.com/patterns/pimpl.html) that we use throughout the project. I personally use this idiom pretty much everywhere and organization-wise too, this is not the first time this problem occurs in our projects. However, I think that the benefits overcome the drawbacks. 

The other workaround, in my opinion even worse but it depends how many time is necessary, is passing a pointer to the public class to the `Impl` constructor.

FirefoxMetzger: Thanks for the added detail 💯 I'll go ahead with the signature you suggested then.

FirefoxMetzger:
I've added the two functions we discussed and made `World::insertModel` a shim over `World::InsertModelFromFile`.

diegoferigo (approved):
Great, thanks for iterating! Merging.

Issue #314, opened 2021-03-31
Enable self-collision for Panda


FirefoxMetzger:
I don't know if this issue should live here or in the `gym_ignition_models` repo. 

Currently, Panda doesn't seem to have self-collisions enabled. Collision with other objects works fine. See for example this video (at around 0:08 - 0:09):

https://user-images.githubusercontent.com/4402489/113131592-43b49580-91d2-11eb-863c-84bcdb292fee.mp4

Would it be possible to enable it by default, or provide an option to enable it?

diegoferigo:
Contacts and collisions are part of what I call the bottleneck features. When enabled, simulations could slow down considerably.

For this reason, contact detection and self collisions are disabled by default ([contacts](https://github.com/robotology/gym-ignition/blob/bba62416e97fe5788019e32cae66355bcb9fe70a/cpp/scenario/gazebo/src/Link.cpp#L168-L171) [1], [collisions](https://github.com/robotology/gym-ignition/blob/bba62416e97fe5788019e32cae66355bcb9fe70a/cpp/scenario/gazebo/src/Model.cpp#L175-L178)). You can enable self collisions model-wise with the following method:

https://github.com/robotology/gym-ignition/blob/bba62416e97fe5788019e32cae66355bcb9fe70a/cpp/scenario/gazebo/include/scenario/gazebo/Model.h#L174-L183

[1] I actually realized that contacts are enabled by default. We should understand what is the impact to the real time factor, and if it makes sense to disable them by default.

FirefoxMetzger:
In python this will need to be cast `to_gazebo()` first:

```
panda.to_gazebo().enable_self_collisions(True)
```

I should first read the docs more thoroughly before opening new issues 🥇 I totally forgot to check the Gazebo API and just looked at the core API. Closing...

diegoferigo:
No worries, we can consider issues as part of a project's documentation :wink: One nice thing with SWIG (much more difficult to get with pybind11) is that the C++ doxygen documentation is copied to the Python's docstring. I admit that, as you, I rarely rely on the online documentation. Nowadays autocompletion is much faster.

PR #317, opened 2021-03-31
Disable contact detection by default

diegoferigo:
As discussed in #314, I was thinking that contact detection was disabled by default. This PR disables them.

The recommended way to use contact is enabling their detection either on a link level, or model-wise (less efficient).

https://github.com/robotology/gym-ignition/blob/bba62416e97fe5788019e32cae66355bcb9fe70a/cpp/scenario/core/include/scenario/core/Link.h#L153-L159

https://github.com/robotology/gym-ignition/blob/bba62416e97fe5788019e32cae66355bcb9fe70a/cpp/scenario/core/include/scenario/core/Model.h#L200-L207

### DIFF
--- a/cpp/scenario/gazebo/src/Link.cpp
+++ b/cpp/scenario/gazebo/src/Link.cpp
@@ -165,7 +165,7 @@ bool Link::createECMResources()
     m_ecm->CreateComponent(m_entity, components::LinearAcceleration());
     m_ecm->CreateComponent(m_entity, components::AngularAcceleration());
 
-    if (!this->enableContactDetection(true)) {
+    if (!this->enableContactDetection(false)) {
         sError << "Failed to enable contact detection" << std::endl;
         return false;
     }
--- a/docs/sphinx/getting_started/manipulation.rst
+++ b/docs/sphinx/getting_started/manipulation.rst
@@ -256,6 +256,12 @@ It shows the following functionalities:
     # Insert the Panda manipulator
     panda = gym_ignition_environments.models.panda.Panda(
         world=world, position=[-0.1, 0, 1.0])
+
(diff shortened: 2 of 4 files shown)


FirefoxMetzger:
Aren't contacts necessary for detecting collision with other objects? I don't know how the DART/ODE handles this, but afaik bullet first checks if objects are in contact, and then checks for collision/self-collision and resolves it appropriately. No contact checking for an object means it will no longer collide with anything in the world neither.

Could we check that this doesn't happen with DART/ODE? I fear this would break the manipulation example otherwise.

diegoferigo:
> Aren't contacts necessary for detecting collision with other objects? I don't know how the DART/ODE handles this, but afaik bullet first checks if objects are in contact, and then checks for collision/self-collision and resolves it appropriately. No contact checking for an object means it will no longer collide with anything in the world neither.

There are two different abstraction in the stack. One thing is the physics, and disabling contact detection there would have the effect you described. Another thing is the simulator, that sends and receives data to / from the physics.

This PR targets the simulator. Disabling contacts, in this level, means that no contact data is extracted from the physics. In other words, `Link::inContact` would always be `false`. However, the physics will always run with full features.

Unfortunately, the implementation of contacts is not optimal in the simulator. The ECM (the graph that contains all the simulation state) stores contact data as `ignition::msgs::Contacts`. If you follow the entire flow of this data from the physics to Python, you'd notice it gets copied much more I'd like.

> Could we check that this doesn't happen with DART/ODE? I fear this would break the manipulation example otherwise.

I already tried running the manipulation example with this PR. The only necessary update is to enable contacts detection explicitly. It makes sense to update the example (and, reminder to myself, also the website) in this PR.

PR #346, opened 2021-05-07
Split `scenario` and `gym-ignition` Python packages

### LINKED ISSUE #344: Split setuptools packaging of `gym-ignition` and `scenario`
It is a while I have this in the back of my mind. For long time now, the pure-Python `gym-ignition` package and the hybrid C++/Python ScenarIO package are completely independent and can happily live in different repositories. However, from the maintenance point of view, this did not and does not really work well for us.

Another option is having two different `setup.py` files, one for `gym-ignition` and another for `scenario`. After all, the existing `gym-ignition` package already installs these two folders inside site-package, and it is not a really common pattern.

From the users point of view, nothing really change. The after-split gym-ignition package will depend on the new scenario package, and if anyone from downstream that has gym-ignition in their setup.py file, they will transitively get also the new scenario package.

From our point of view, instead, we should only do minor modifications to our CI/CD pipeline.

From the superbuild point of view (cc @traversaro) there's nothing to change. The CMake project will not be touched.

diegoferigo:
Closes #344. After this change, it should be more clear that gym-ignition is meant for RL applications, and scenario for a more broad use that is not related to RL. Those interested in just `scenario` can avoid to install also `gym-ignition`.

- Created new package (with `setup.py`) in the `scenario` folder, former `cpp`.
- `gym-ignition` pure Python package now depends from `scenario`.
- Update Docker images (currently only used in the website pipeline).

CI/CD pipeline changes:

- Split workflow in multiple jobs
- The workflow runs on default runners, it will be easier to add multiplatform support in the future
- The colcon workspace (nightly channel) is built only once and it gets cached even if there are build or test errors
- The new `Build and Test` job can be integrated more easily with conda (#279, #313)
- Create `scenario` wheels for both Python 3.8 and 3.9
- New complete artifact archive with both sdist and wheels for `scenario` and `gym-ignition`

Screenshot_20210507_154215

TODO:

- [x] Finalize new `scenario` README that will be included in the new PyPI package
- [x] Update installation instructions
- [x] Clean docker images
- [x] Fix website workflow
- [x] ~Claim the `scenario` PyPI package (https://github.com/pypa/pypi-support/issues/1082)~ Using `scenar-io` while waiting.
- [x] Check if the PyPI token works for both `scenario` and `gym-ignition`, otherwise generate a new one


diegoferigo:
Waiting the decision on the PyPI name in the dedicated issue, maybe we can start moving this forward with the `scenar-io` name. The import will remain the same (`scenario`), it's just the PyPI package that will have a temporarily name. I think until everything remains in `devel` this should be fine.

diegoferigo:
I guess we can move forward in steps, I'd like to reach sooner or later what described in #355 but it can be done in a next PR.

traversaro (approved):

PR #364, opened 2021-06-29
Update setuptools resources introducing `setup.cfg`

### LINKED ISSUE #355: Move content from `setup.py` to `setup.cfg`

diegoferigo:
Closes #355


diegoferigo:
ping @traversaro

traversaro (approved):

PR #368, opened 2021-07-01
Test a multi-world simulation with DART and bullet collision detector

diegoferigo:
Upstream support introduced in https://github.com/ignitionrobotics/ign-gazebo/pull/684. In similar setup, ODE crashes due to https://github.com/ignitionrobotics/ign-gazebo/issues/18, however, replacing ODE with bullet should prevent that. This setup seems a good candidate whether the aim is:

1. A multi-world simulation (advanced feature of Ignition Gazebo).
2. A single-process multi-thread setup where each thread simulates a single world (e.g. in Python's with `threading` or in C++ with `std::thread`).

This test is simple enough that also ODE does not crash, but the simulation slows down considerably when the detector has to deal with contact between the sphere and the ground. Using bullet seems faster in this case.

Note: I also tried a more complex single-world simulation (the panda manipulation example) and bullet with both pgs and dantzig solvers seems working fine. However, compared to the default ODE collision detector, the RTF is much slower (3-5x).

cc @nicholaspalomo

### DIFF
--- a/tests/common/utils.py
+++ b/tests/common/utils.py
@@ -3,6 +3,7 @@
 # GNU Lesser General Public License v2.1 or any later version.
 
 import pytest
+import dataclasses
 from typing import Tuple
 import gym_ignition_models
 from gym_ignition.utils import misc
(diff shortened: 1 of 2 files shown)


traversaro (approved):

diegoferigo:
Note that this test changes the collision detector through the world's SDF. Similarly to what we do for the physics engine, it would be nice having APIs that simplify this setting. However, it is not part of this PR and it might be implemented in the future in a new PR. For the moment, using SDF for testing purposes is more than enough.

diegoferigo:
Unfortunately this test is flaky and CI didn't catch it before merging. Updated in 027f0a4.

PR #370, opened 2021-07-02
Run the Python tests in different processes

diegoferigo:
This should prevent flaky tests that might segfault to crash the all the test suite. It uses [`pytest-forked`](https://github.com/pytest-dev/pytest-forked) plugin.

Note that this plugin does not work in Windows, but we do not yet support this platform.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,6 +73,7 @@ where = python
 testing =
     pytest
     pytest-xvfb
+    pytest-forked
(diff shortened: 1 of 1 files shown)


traversaro (approved):

diegoferigo:
For the time being, since running in different processes didn't solve the occasional segfault of tests, we disabled the test in #374.

Issue #289, opened 2021-02-03
Disable the automatic loading of Gazebo Server plugins


diegoferigo:
After https://github.com/ignitionrobotics/ign-gazebo/pull/281, if the SDF world file does not specify any plugin (such as physics, scene broadcaster, etc), the server will load a default set of plugins specified in a new `server.config` file.

This likely will alter the situation we were expecting before. In fact, we offer and use programmatic APIs to load world and model plugins. We don't expect any plugin to be loaded by default.

I still have to update my source installation (the PR was merged into dome with https://github.com/ignitionrobotics/ign-gazebo/pull/602 just yesterday). I hope that by exporting an empty `IGN_GAZEBO_SERVER_CONFIG_PATH` as documented [here](https://github.com/ignitionrobotics/ign-gazebo/blob/0f284ba7f14863e449cdbbe9fae821f5bd31d445/include/ignition/gazebo/ServerConfig.hh#L406-L426) would suffice.

diegoferigo:
At first look, the `server.config` file is read when no plugins are specified in the following locations:

- SDF world file
- `ServerConfig` object passed to `Server`

Waiting to address , we rely on a workaround involving a singleton (-.-) to get the full simulation state, implemented as a plugin loaded through `ServerConfig`:

https://github.com/robotology/gym-ignition/blob/50caccfa1b5ddb617ec0b1aea98bfec72c8fbadf/cpp/scenario/gazebo/src/GazeboSimulator.cpp#L648-L652

I leave this issue open as reminder. Until we keep this workaround, no action is necessary. [Exporting](https://github.com/robotology/gym-ignition/blob/50caccfa1b5ddb617ec0b1aea98bfec72c8fbadf/bindings/__init__.py#L23-L42) an empty `IGN_GAZEBO_SERVER_CONFIG_PATH` [as documented](https://github.com/ignitionrobotics/ign-gazebo/blob/0f284ba7f14863e449cdbbe9fae821f5bd31d445/include/ignition/gazebo/ServerConfig.hh#L406-L426) would not harm, however now is is not yet strictly needed in our current situation.

diegoferigo:
Closed via https://github.com/robotology/gym-ignition/pull/379.

PR #379, opened 2021-08-02
Remove the need to use a singleton to get the ECM pointer

### LINKED ISSUE #378: Simplify logic to get the Entity Component Manager and Event Manager
In Ignition Gazebo, the [`Server`](https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1/include/ignition/gazebo/Server.hh) class contains one or multiple [`SimulationRunner`](https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1/src/SimulationRunner.hh) objects (in most cases, just one), each simulating an independent world. The data storage of each `SimulationRunner` is implemented in a Entity Component System.

Historically, the ECS has always been hidden from the user (https://github.com/ignitionrobotics/ign-gazebo/issues/248). Though, the entire Gazebo ScenarIO logic is based on the fact that we can access and alter this data storage. Until now, we had to rely on a kind of dirty workaround: a plugin that implements only the [`ECMProvider::Configure`](https://github.com/robotology/gym-ignition/blob/311b08ef16/cpp/scenario/plugins/ECMProvider/ECMProvider.cpp) callback that gets the pointer to the ECM (and the event manager) and stores it in a singleton. Then, within the same process, the singleton could be used to get these pointers and consume them.

It's easy to see that the combination of plugins + singletons + python could create problems. In fact, we struggled with some hairy issues in the past (see a recap in #354 that, even if not specific to this, is highly related).

Luckily, the new upstream https://github.com/ignitionrobotics/ign-gazebo/pull/936 allows us to gather the ECM without the need of using the plugin + singleton :tada: If I understood, the `ECMProvider` class of the plugin can be allocated together with the `Server` (e.g. in the `GazeboSimulator` class) and its system directly added. Then, the ECM and event manager can be gathered from the `ECMProvider` object with new getters, and stored, if needed, in our `GazeboSimulator` class.

Likely this solves for good the need to alter the dlopen flags as described in #7 (very old issue!). Though, if it does not harm anyone, I'd leave this workaround there in any case.

diegoferigo:
Closes #378 (thanks to https://github.com/ignitionrobotics/ign-gazebo/pull/936, further details in https://github.com/ignitionrobotics/ign-gazebo/issues/248) and implements #289.

Requires https://github.com/ignitionrobotics/ign-gazebo/pull/953. Merged now in upstream, waiting forward porting to Edifice.


scenario/src/gazebo/src/GazeboSimulator.cpp
```diff
@@ -664,26 +687,52 @@ std::shared_ptr<ignition::gazebo::Server> GazeboSimulator::Impl::getServer()
...
+            // Get the ECM and EventManager pointers
+            detail::SimulationResources resources;
+            resources.ecm = provider->ecm;
+            resources.eventMgr = provider->eventMgr;
+            this->resources[provider->worldName] = resources;
+        }
+
         sDebug << "Starting the gazebo server" << std::endl;
 
+        // TODO: is this redundant now?
```
diegoferigo, line 735: Reminder: address this before merging

traversaro (approved):
Approving, even if I guess you want to wait for a release of the required changes to merge this.

diegoferigo:
> Approving, even if I guess you want to wait for a release of the required changes to merge this.

Yes, I need https://github.com/ignitionrobotics/ign-gazebo/pull/953 to be forward ported to upstream's ign-gazebo5 (Edifice) in order to merge the PR to `devel`, and a new Ignition Gazebo minor release to merge `devel` in `master`.

diegoferigo:
Forward porting has been done in https://github.com/ignitionrobotics/ign-gazebo/pull/968. CI is happy, at least to target `devel`, merging.

PR #383, opened 2021-09-10
Bump Ignition distribution to Fortress

diegoferigo:
- [Roadmap](https://ignitionrobotics.org/docs/all/roadmap#fortress)

---

- Update CMake logic
  - [x] Add the new version in [`FindIgnitionDistribution.cmake`](https://github.com/robotology/gym-ignition/blob/3b1787912ed8a86d93ce1cfbe9d72c53380468f6/cmake/FindIgnitionDistribution.cmake#L9)
  - [x] Create a new [`ImportTargets<>.cmake`](https://github.com/robotology/gym-ignition/blob/master/cmake/ImportTargetsDome.cmake)
  - [x] Update [`SUPPORTED_IGNITION_DISTRIBUTIONS`](https://github.com/robotology/gym-ignition/blob/3b1787912ed8a86d93ce1cfbe9d72c53380468f6/CMakeLists.txt#L149)
- Update CI to:
  - [x] Build and test gym-ignition against Fortress
  - [x] Build and upload wheels by default against Fortress 
  - [x] Build docker images with Fortress 
- [x] Check that all documentation is updated
- [x] Check that building the wheel locally succeeds by passing custom Dome flag to pip / setup.py (even if not supported officially)
- [ ] Mark the new workflow `Build and Test (<New Ignition Distribution>)` check as _Required_ in the branches setting

PR #387, opened 2021-09-14
Fix `test_velocity_direct`

diegoferigo:
Not sure why now two steps are necessary. Maybe related to the recent enforcing of the joint velocity limits.


traversaro (approved):

diegoferigo:
Maybe I was updating the limits in the wrong place. I tried again in #383 and it worked, no need to perform two steps.

PR #391, opened 2021-09-28
Application of external wrenches: new `empty` method and new Python test

diegoferigo:
- Adds new test for checking if wrenches are applied for the right time interval.
- Adds a new `LinkWrenchCmd::empty` to check if there are wrenches to apply. This will be used by the Physics system in a new PR to prevent applying a zero wrench each time.

### DIFF
--- a/tests/test_scenario/test_external_wrench.py
+++ b/tests/test_scenario/test_external_wrench.py
@@ -1,0 +1,56 @@
+# Copyright (C) 2021 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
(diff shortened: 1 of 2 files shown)


tests/test_scenario/test_external_wrench.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # Run the simulation for a while
+    for _ in range(500):
+        gazebo.run()
+        assert link.position()[2] == pytest.approx(initial_height)
+
+    # Run the simulation for a while
+    for _ in range(500):
+        gazebo.run()
+        print(link.position()[2])
```
diegoferigo, line 56: leftover

diegoferigo:
The Physics system addition will be included in #392.

